PR #227, opened 2025-01-15
Update linux_build.sh

birwin88:
Trying to add support for Fedora 41.  

Fedora 41 makes it through without errors now.  

Have not been able to test on a Fedora 40 or below system..


ClassicOldSong:
LGTM

PR #465, opened 2025-03-06
Adding template image for applications

ryanbbernstein:
## Description
I wanted the applications I add to have a similar look to the existing ones so i spent some time trying to match the gradient. Figured it could be useful for others as well.

### Screenshot
template

### Issues Fixed or Closed
None


ClassicOldSong:
Thanks for your contribution but this is not needed in many clients. When the image is not present, they'll display the app's name instead, but with an empty image users won't see app names directly, making it hard to distinguish between apps.

ryanbbernstein:
This is meant as a template for people adding their own icon on top (in external image editor) and not for use as is

ClassicOldSong:
Ah ok, maybe we can make a built-in generator for it.

ryanbbernstein:
I was thinking something like that would be nice. Simple web page to slap a logo on the template. I can get something like that together and send a PR.

PR #203, opened 2025-01-06
Rename pin to pairing 

fabienheureux:
Following https://github.com/ClassicOldSong/Apollo/discussions/192 
We agreed on renaming Pin to Pairing 

## TODO 

- [ ] Translations
  - [ ] de
  - [ ] es
  - [ ] it
  - [ ] ja
  - [ ] pt
  - [ ] ru
  - [ ] sv
  - [ ] tr
  - [ ] zh

PR #478, opened 2025-03-10
Improves pt-br translations

bestknighter:
The previous translation of pt-br texts were mechanical and got really far, but not far enough. I'm a native pt-br speaker and did some tweaks to achieve a version that is more correct, natural and compliant with officially translated technical jargon.


bestknighter:
If you want to, I can address #203 and add the "pairing" entry. So, there they'll only need to remove the old "PIN" one.

If not, the correct translation would be "Pareamento". I'm leaving it here for future references if those changes ends up being done and merged some day.

ClassicOldSong:
Thanks! There're some Apollo specific parts added only to the English and Chinese locale, if you can would you translate them as well?

bestknighter:
Oh, I see. Sure! I'll let you know when I'm done with it.

bestknighter:
Done! I also did some fixes to the en.json. It might be a good idea to double check other locale files.

ClassicOldSong:
Ah, thank you very much 
! Didn't noticed I got those typos!

I'll check later after I get up 😂

bestknighter:
No worries. Have a nice rest!

ClassicOldSong:
Looks great! Thank you for improving the translation!

BrunoVieiraL:
@bestknighter Mind if i make some changes? There's some incorrect translations yet like "Residencial" to indicate the "Home" and some other changes. If those files were on Weblate i could do it on my phone as well but since it's only through PR , i will take some time to translate anything. And thanks for the job you've done.

bestknighter:
@BrunoVieiraL Sure! Go ahead. There's a lot of strings, I was suspicious that I haven't missed anything lol

PR #604, opened 2025-04-10
Minor Windows Compile / Packaging / Run issues when compiling from scratch

thesystemcoder168:
Minor Windows Compile / Packaging / Run issues when compiling from scratch

Background: Building locally on Windows 10

1. FILE CHANGED - CMakeLists.txt
Change to add additional INCLUDE path to get rid of the fatal error: nlohmann/json.hpp: No such file or directory

This occurs in the Windows compile for files like:
dxgi.cpp and audio.cpp

Example:
In file included from C:/compiler-sunshine/msys64/home/vm/Apollo/tools/dxgi.cpp:6:
C:/compiler-sunshine/msys64/home/vm/Apollo/src/utility.h:20:10: fatal error: nlohmann/json.hpp: No such file or directory

2. FILE CHANGED - building.md
Pull from the https://github.com/ClassicOldSong/Apollo.git instead of lizardbyte one and change the directory to apollo instead of sunshine.

3. FILES ADDED - SudoVDA.dll and nefconc.exe
When the .ZIP and INSTALLER .EXE packaging are created from a fresh build, they can be used on a CLEAN system and have all of the dependencies so that the SudoVDA driver will work after installation.

Purpose:
Going to propose other Windows changes and do not want to be caught up in Windows compiling/packaging issues.


thesystemcoder168:
1.
>There should be no issue in cmake if you have installed nholmann/json from pacman 
If I run pacman manually, by adding it and I got rid of the CMakeList.txt changes, then it looks like the compile works.
what I used: pacman -S "mingw-w64-ucrt-x86_64-nlohmann_json"

So it looks like the Building.md file is missing the mingw-w64-ucrt-x86_64-nlohmann_json dependency.

2. 
>I have kept the driver dll and nefcon away explicitly as binaries may not compatible with GPL license, and driver signature may expire afterwards. It's better to compile the driver fresh yourself.

I will see if I can get rid of those binary adds. 
As for immediate driver files situation, would it be worth it to add something in the build documentation to explicitly state that those files are not present? (To prevent confusion)

As for the bigger driver situation, that topic would probably be more of a discussion forum topic as compiling / signing / etc is not a simple issue.

//////////////////

Summary:
1. I will try to get rid of the CMakeList.txt change and try to get rid of the binary files. I will also add the "mingw-w64-ucrt-x86_64-nlohmann_json" into the packages that pacman installs in the build document.

thesystemcoder168:
Now, there is only 1 file changed. It is the build documentation.

A. The correct git command and directory. (apollo)
B. mingw-w64-ucrt-x86_64-nlohmann_json was added to the pacman installation list.

ClassicOldSong:
LGTM, thanks again for the contribution!

PR #606, opened 2025-04-11
Added better Images for the UI

josf0:
Hi there. This is my first time contributing to a FOSS project so please excuse me if I've done something stupid with the repo and feel free to let me know about it.

I felt like the box images in the UI were a bit outdated so I've personally made some and added it to the assets directory. I'm open to revise them if you have any preference.

Happy to contribute and thanks a lot for making this project it has made my game streaming so effortless, keep going!


josf0:
Happy to contribute! 

Ya you're right about it, I'll try to make a separate repository. 

Also could you please provide on how I can compile this source code into an executable file and also I'm trying to tweak Artemis UI too so if you don't mind how can I compile that too?

ClassicOldSong:
Artemis's UI is decided by the client, not by the host.

Both Artemis and Apollo's UI will undergo a complete rewrite in the future, so contributing on improving the current one will be a waste of time. It's not that bad and functional, just don't break it before the refactor😂

josf0:
Okay thanks for the info.

PR #743, opened 2025-05-20
Supporting MSVC [WIP]

Luiz-Monad:
What's the deal with MinHook ? 
I changed it so it downloads automatically like the json library.

### DIFF
--- a/cmake/dependencies/common.cmake
+++ b/cmake/dependencies/common.cmake
@@ -19,6 +19,16 @@ add_subdirectory("${CMAKE_SOURCE_DIR}/third-party/libdisplaydevice")
 include("${CMAKE_MODULE_PATH}/dependencies/nlohmann_json.cmake")
 find_package(OpenSSL REQUIRED)
 find_package(Threads REQUIRED)
+find_package(DEPcurl REQUIRED)
+
+# miniupnp
+find_package(DEPminiupnpc REQUIRED)
+
+# use the ffmpeg compiled by the MSVC from the VCPKG
+if(DEFINED CMAKE_TOOLCHAIN_FILE)
+    find_package(ffmpeg REQUIRED)
+    set(FFMPEG_PREPARED_BINARIES ON)
+endif()
 
 # ffmpeg pre-compiled binaries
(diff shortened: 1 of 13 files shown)


ClassicOldSong:
Have you tried installing minhook using pacman first?

cmake/dependencies/common.cmake
ClassicOldSong, line 1: libcurl and other dependencies are also required on Windows, why removing this part?

Luiz-Monad: I can't use pacman on Windows, so I thought that was the solution for the missing dependency. 
That was a mistake, I have to provide the libcurl. I thought they weren't being linked. Either way, pkg_check_modules is not going to work on Windows, unless under the vc_pkg as it provides an abstraction layer. So that's why I'm asking how to proceed.
The thing is that I'm not using the msys environment to build the software, so I might be having problems with implicit dependencies that comes for it.

Luiz-Monad: I think I understood what I was doing wrong. I'm going to provide a way of also getting all the other dependencies, so the project can be truly CMake multiplatform without depending on the msys environment. Hold on.

ClassicOldSong:
This could be big as some parts of the code have specific branches for MSYS2. I recommend keeping the MSYS2 environment as it's more unified cross platform.

ClassicOldSong:
Ans GCC has advantages in performance and compiler features/extensions.

Luiz-Monad:
> This could be big as some parts of the code have specific branches for MSYS2. I recommend keeping the MSYS2 environment as it's more unified cross platform.

I was able to make the Apollo project itself compile on CMake targeting the MSVC.
There's the customized FFMpeg on LizardByte-infrastructure, but for that I can take the vcpkg ffmpeg and apply the patches on top of it, Its missing the SVT-AV1 thou
I might just use FFmpeg as a DLL. 

It was to my surprise that the vcpkg had the portfile for MinHook for Windows. So I don't need to change much of this project to be able to do what I needed. 

The dependencies of this project are bog standard, other than the FFMpeg, which still uses autotools (eew).
It was a fun experiment.

Issue #676, opened 2025-04-29
Windows graphics capture 60 fps limit?


bdonutd:
Hi, so I'm trying to use the windows graphics capture (wgc) api with apollo services turned off and running 3 apollo.exe with admin rights streaming to a Quest 3 vr headset with 3 Artemis clients using warp drive 2 with resolution at 1440p, 100mb bitrate and 120 fps. It works great but the streams are limited to 60 fps for some reason??? Is this a problem in the source code or am I doing something wrong with the setup in apollo? setting the capture setting to desktop duplication api works at 120 fps with all the settings the same in the clients, but I need wgc because I need to be to see the xbox game bar overlay, which I can't get working with the desktop duplication caputre method. 

Any help or fix would be greatly appreciated!

ClassicOldSong:
Dup: https://github.com/ClassicOldSong/Apollo/issues/620

Also starting multiple clients can still remain SYSTEM privilege and you can still use DXGI for capturing. Checkout https://github.com/ClassicOldSong/Apollo/wiki/How-to-start-multiple-instances-of-Apollo

Issue #620, opened 2025-04-15
[Feature Request] WGC Improvement


merlinch:
WGC (Windows.Graphic.Capture) have several issues that need to be addressed on

1. Could only send 60FPS to Artemis (even if it set to 120fps or custom)
2. Could only achieved when Apollo/Sunshine isn't running as services (manually go to exe file)

i was about to write the third one with "Display from 60hz to 120hz only sending max 48 fps from Apollo to Artemis" but it seems this is fixed, might need further testing. I know this feature is still beta, but improvement is improvement

ClassicOldSong:
It's a Windows session problem that image can't be captured when running as SYSTEM, there might be no easy fix for this.

The FPS problem needs some investigations.

roob-p:
@merlinch @ClassicOldSong I've made a utility for Sunshine and Apollo that automates switching between DDX and WGC: when the PC is at the lock screen, it switches to DDX, and when the computer is unlocked, it automatically switches to WGC (although the switch requires reconnecting from the client).

I also added an UAC-handling system for WGC: Sunshine/Apollo temporarily switches to DDX when a UAC prompt is detected, and returns to WGC once the prompt is closed (this requires two reconnections from the client in total). The behavior of the UAC-handling system, as well as other aspects, can be customized. 

The WGC script can also stay active in the system tray to monitor future UAC prompts and both the WGC and DDX scripts can show an icon in the system tray indicating the current capture method.

You can check it out here if you want: https://github.com/roob-p/SunshineCaptureSwitcher

PR #785, opened 2025-05-30
Resolve 60 fps lock issue for WGC capture by setting the MinUpdateInterval property of the WGC capture session.

danielmartina:
Resolve 60 fps lock issue for WGC capture by setting the MinUpdateInterval property of the WGC capture session.

Fixes: #676 and partially #620

### DIFF
--- a/src/platform/windows/display_wgc.cpp
+++ b/src/platform/windows/display_wgc.cpp
@@ -139,7 +139,7 @@ namespace platf::dxgi {
     }
     try {
       if (winrt::ApiInformation::IsPropertyPresent(L"Windows.Graphics.Capture.GraphicsCaptureSession", L"MinUpdateInterval")) {
-        capture_session.MinUpdateInterval(winrt::TimeSpan{ 10000 });
+        capture_session.MinUpdateInterval(winrt::TimeSpan{ 10000000 / (config.framerate * 2) });
       }
       else {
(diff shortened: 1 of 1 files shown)


danielmartina:
I do not see any frame pacing issues when looking at TestUfo for example. On the topic of wasted frames, it is hard to say as I do not see into the inner workings of the WGC API.

10000 corresponds to a framerate of 1000 Hz, which is a lot.

ClassicOldSong:
https://github.com/ClassicOldSong/Apollo/blob/d293fee801be14d07129e561cbe30e8a12376bfb/src/platform/windows/display_wgc.cpp#L268

Can you check the time difference? A rough implemtation is enough, set a static variable here and diff it each time when snapshot is called

ClassicOldSong:
Tried on my computer, but with or without this patch WGC never goes over 24fps and IDK why...

ClassicOldSong:
It's AppleMusic... Something strange happens if I have AppleMusic running and then connect the virtual display, it stutters like hell.

Close it and then open it within the virtual display, it's normal again. But then I disconnects the virtual display, AppleMusic stutters on the physical monitor......

How weird it is.

ClassicOldSong:
Tested working fine on my machine, even with the weird refresh rates.

Thanks for the contribution!

PR #789, opened 2025-05-31
Update Boost hash to version 1.88.0

nil-andreas:
Updating the hash of Boost version 1.88.0. Verified against https://github.com/boostorg/boost/releases/download/boost-1.88.0/boost-1.88.0-cmake.tar.xz.txt.

Without this change FetchContent will fail to verify the downloaded boost-1.88.0 and the build will fail.


ClassicOldSong:
Thanks, but I think we should prefer system installed boost instead of building it from source.

Issue #791, opened 2025-05-31
Building on Linux broken after introducing Boost v.1.88.0


nil-andreas:
OS: NixOS

How to reproduce:
1. `git checkout master`
2. `nix-shell --pure`
3. `mkdir -p build && cd build`
4. `cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release -DSUNSHINE_ENABLE_WAYLAND=ON -DSUNSHINE_ENABLE_X11=ON-DSUNSHINE_ENABLE_DRM=ON -DBUILD_TESTS=OFF -DBUILD_DOCS=OFF -DSUNSHINE_BUILD_APPIMAGE=OFF -DSUNSHINE_ENABLE_CUDA=OFF`
5. `ninja`

I've attached logs for [cmake](https://github.com/user-attachments/files/20535862/cmake_log.txt) and [ninja](https://github.com/user-attachments/files/20535863/ninja_log.txt). 

If I instead `git checkout v0.3.5-alpha.5` (which uses Boost v.1.87.0) the above works

I'm using the following shell.nix for reproducibility. If you want to do the same you need to install nix and save the file in Apollo folder as shell.nix.

```nix
# shell.nix for reproducing Apollo build error with Boost 1.88.0
{pkgs ? import (fetchTarball "https://github.com/NixOS/nixpkgs/archive/96ec055edbe5ee227f28cdbc3f1ddf1df5965102.tar.gz") {}}:
pkgs.mkShell {
  name = "apollo-boost188-error-repro";

  nativeBuildInputs = [
    pkgs.cmake
    pkgs.pkg-config
    pkgs.python3
    pkgs.makeWrapper
    pkgs.wayland-scanner
    pkgs.autoPatchelfHook
    pkgs.ninja
  ];

  buildInputs =
    [
      pkgs.avahi
      pkgs.libevdev
      pkgs.libpulseaudio
      pkgs.xorg.libX11
      pkgs.xorg.libxcb
      pkgs.xorg.libXfixes
      pkgs.xorg.libXrandr
      pkgs.xorg.libXtst
      pkgs.xorg.libXi
      pkgs.openssl
      pkgs.libopus
      pkgs.libdrm
      pkgs.wayland
      pkgs.libffi
      pkgs.libcap
      pkgs.curl
      pkgs.cacert
      pkgs.pcre
      pkgs.pcre2
      pkgs.libuuid
      pkgs.libselinux
      pkgs.libsepol
[...]

PR #792, opened 2025-05-31
Fix build problems on Linux

nil-andreas:
This PR fixes the problem described in https://github.com/ClassicOldSong/Apollo/issues/791


ClassicOldSong:
LGTM

PR #886, opened 2025-06-25
Feature: Added option to disable rumble on controllers from the Apollo side

thesystemcoder168:
Feature: Added option to disable rumble on controllers from the Apollo side

Use case:
Some users do not always like the rumble functionality and it can interfere with their game play. Disabling the rumble function on a case-by-case system/phone/game/etc basis can be difficult especially for non-technical users or trying to quickly set things up.

Feature:
In the Configuration->Input tab, there is the option to block the rumble functionality. The default is to not disable this functionality (i.e. false). Essentially nothing is done in the rumble() function when it is blocked (i.e. enabled/true).

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -572,6 +572,7 @@ namespace config {
     true,  // high resolution scrolling
     true,  // native pen/touch support
     false, // enable input only mode
+    false, // enable_block_rumble_messages_to_controllers
   };
 
   sunshine_t sunshine {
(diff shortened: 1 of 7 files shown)


ClassicOldSong:
I think we might better inverting the toggle, making it "Enable Rumble" and make it enabled by default... The reversed logic is a pain to deal with.

thesystemcoder168:
Changed the logic to enable/disable rumble functionality.

PR #1038, opened 2025-08-22
Windows Graphics Capture in Service

Nonary:
This allows Apollo to be able to capture the screen using Windows.Graphics.Capture API even while it is running as System.

WGC has roughly the same performance as DXGI and does NOT freeze when VRAM is exceeded even with HAGS enabled.

### DIFF
--- a/src/platform/windows/display_vram.cpp
+++ b/src/platform/windows/display_vram.cpp
@@ -7,7 +7,7 @@
 
 // platform includes
 #include <d3dcompiler.h>
-#include <DirectXMath.h>
+#include <directxmath.h>
 
 extern "C" {
 #include <libavcodec/avcodec.h>
--- a/src/platform/windows/input.cpp
+++ b/src/platform/windows/input.cpp
@@ -3,6 +3,7 @@
  * @brief Definitions for input handling on Windows.
  */
 #define WINVER 0x0A00
+
 // platform includes
 #include <windows.h>
 
--- a/src/platform/windows/ipc/pipes.cpp
+++ b/src/platform/windows/ipc/pipes.cpp
@@ -27,7 +27,7 @@
 #include <AclAPI.h>
 #include <combaseapi.h>
 #include <sddl.h>
-#include <Windows.h>
+#include <windows.h>
 #include <winrt/base.h>
 
 // local includes
--- a/src/platform/windows/misc.h
+++ b/src/platform/windows/misc.h
@@ -11,7 +11,7 @@
 #include <system_error>
 
 // platform includes
-#include <Windows.h>
+#include <windows.h>
 #include <winnt.h>
 
 namespace platf {
--- a/tools/apollo_wgc_capture.cpp
+++ b/tools/apollo_wgc_capture.cpp
@@ -29,15 +29,15 @@
 // platform includes
 #include <d3d11.h>
 #include <dxgi1_2.h>
-#include <inspectable.h>  // For IInspectable
-#include <ShellScalingApi.h>  // For DPI awareness
-#include <Windows.Graphics.Capture.Interop.h>
-#include <Windows.h>
+#include <inspectable.h>  // for iinspectable
+#include <shellscalingapi.h>  // for dpi awareness
+#include <windows.graphics.capture.interop.h>
+#include <windows.h>
 #include <winrt/base.h>
-#include <winrt/Windows.Foundation.h>
-#include <winrt/Windows.Foundation.Metadata.h>  // For ApiInformation
-#include <winrt/Windows.Graphics.Capture.h>
-#include <winrt/Windows.Graphics.DirectX.Direct3D11.h>
+#include <winrt/windows.foundation.h>
+#include <winrt/windows.foundation.metadata.h>  // for apiinformation
+#include <winrt/windows.graphics.capture.h>
+#include <winrt/windows.graphics.directx.direct3d11.h>
 
 // Manual declaration for CreateDirect3D11DeviceFromDXGIDevice if missing
 extern "C" {


Nonary:
This still needs some minor editing to remove pointless changes, so leaving it in draft for now but will clean up some of the rough edges in a bit

xanderfrangos:
Great to see this PR moving over to Apollo! I was worried all of your work would end up stranded in yet another Sunshine fork. It seemed to work well when I tested the Sunshine PR. Hope you're able to work with ClassicOldSong on getting it integrated here.

Nonary:
> Great to see this PR moving over to Apollo! I was worried all of your work would end up stranded in yet another Sunshine fork. It seemed to work well when I tested the Sunshine PR. Hope you're able to work with ClassicOldSong on getting it integrated here.

Actually, my plan is to start a new fork... but my plan is to support both Apollo and Sunshine.

Backported features like WGC will be sent directly to Apollo if she chooses to accept them.

xanderfrangos:
Well the fact that you're sharing this with Apollo alleviates the concern of further fracturing the features across forks. I didn't mean to diss your fork. 😅

I've been testing the PR since yesterday, and it's been stable for me. The WGC Constant option works great for remote desktop use, and WGC Variable is working as a drop-in replacement for DDX (but I haven't tested it a ton). It'd be amazing if there was a way to switch WGC modes per profile. 

Fingers crossed ClassicOldSong accepts the PR. 🤞

BTW, I think the `rowspan` needs to be set to 7 here: https://github.com/Nonary/Vibepollo/blob/wgc/simplified/docs/configuration.md?plain=1#L1991

ClassicOldSong:
I will merge but I need to review and make sure it won't conflict with other parts of Apollo.

Me and @Nonary actually discussed this further and have more wild thoughts, but I'll make sure this pr gets merged as a stepping stone for our further explorations.

Nonary:
I don't mind back-porting the features its just quite a lot, I am already diverged off with 30,000 changes. 
Some of these Apollo already does 

I have done so far:

1. Completely reworked frontend with full mobile support and more modern looking feel.
2. Reworked authentication to be session based and works without redirects.
3. API token support with scopes
4. Reworked the display swap API to be more resilient against 24H2 and added fallbacks so that it can swap even when resolution don't exist (which fixes a lot of the issues of the current implementation). Trying to make this work even if user reboots or logs out before the display swap is completed it should make sure the computer logs back in with their old configuration instead of being stuck on a virtual display [still working on that]
5. Complete Playnite Integration, auto sync all games from playnite and easily add them with one click on frontend. This includes being able to auto close sessions when game closes and support termination from client side (Moonlight/Artemis).
6. Revamped configuration to be hot-loadable, you can change capture type and nearly all settings without having to restart sunshine.
7. Windows Graphics Capture working in service mode


Planning to also integrate RTSS frame limiting as well, but have not started on that yet.

Basically I have a lot of ideas that I will be working on and needed my own fork for now because it will take many weeks or even months to get those all reviewed and polished up for back-porting.


I just figured the more high value ones like WGC get back-ported first since mostly everyone would benefit from that where as Playnite for example is more niche.

ClassicOldSong:
> Trying to make this work even if user reboots or logs out before the display swap is completed it should make sure the computer logs back in with their old configuration instead of being stuck on a virtual display [still working on that]

Actually with Apollo stucking in a virtual display will never happen. SudoVDA only activates on request and requires a constant ping to keep alive, so even if Apollo crashes the virtual display won't lock users out.

I don't recommend using third party apps to manage display arrangements, since they'll always conflict with Windows's native implemtation. The way Apollo creates and removes a display just takes advantage of Windows's native behavior, so it's always safe to use.

Nonary:
> > Trying to make this work even if user reboots or logs out before the display swap is completed it should make sure the computer logs back in with their old configuration instead of being stuck on a virtual display [still working on that]
> 
> Actually with Apollo stucking in a virtual display will never happen. SudoVDA only activates on request and requires a constant ping to keep alive, so even if Apollo crashes the virtual display won't lock users out.

There is something weird though starting since 24H2 where if for example the main monitor is a TV set and it goes to a deep sleep, the profile restores just doesn't work properly. A lot of external display swap tools also broke mysteriously in 24H2 and still kind of investigating why right now.

> I don't recommend using third party apps to manage display arrangements, since they'll always conflict with Windows's native implemtation. The way Apollo creates and removes a display just takes advantage of Windows's native behavior, so it's always safe to use.

I won't be using a third party app, I will be using the Windows API directly to fix the display swap code basically.

ClassicOldSong:
> > > Trying to make this work even if user reboots or logs out before the display swap is completed it should make sure the computer logs back in with their old configuration instead of being stuck on a virtual display [still working on that]
> > 
> > 
> > Actually with Apollo stucking in a virtual display will never happen. SudoVDA only activates on request and requires a constant ping to keep alive, so even if Apollo crashes the virtual display won't lock users out.
> 
> There is something weird though starting since 24H2 where if for example the main monitor is a TV set and it goes to a deep sleep, the profile restores just doesn't work properly. A lot of external display swap tools also broke mysteriously in 24H2 and still kind of investigating why right now.

Apollo actually don't need profile swapping. It removes the virtual display upon stream termination, and Windows restores to its own saved state. So there's totally no need for Apollo to manage a separate state.

xanderfrangos:
> Planning to also integrate RTSS frame limiting as well, but have not started on that yet.

FYI, I asked Unwinder about getting LimitDenominator supported using the official SDK, and it'll be added in the next RTSS beta: https://forums.guru3d.com/threads/limitdenominator-support-in-rtsshooks-dll.457441/#post-6355395

So if you want to use that instead of your Powershell scripts, you can. 

(Sorry for the off-topic discussion!)

PieroPontra:
> Well the fact that you're sharing this with Apollo alleviates the concern of further fracturing the features across forks. I didn't mean to diss your fork. 😅
> 
> I've been testing the PR since yesterday, and it's been stable for me. The WGC Constant option works great for remote desktop use, and WGC Variable is working as a drop-in replacement for DDX (but I haven't tested it a ton). It'd be amazing if there was a way to switch WGC modes per profile.
> 
> Fingers crossed ClassicOldSong accepts the PR. 🤞
> 
> BTW, I think the `rowspan` needs to be set to 7 here: https://github.com/Nonary/Vibepollo/blob/wgc/simplified/docs/configuration.md?plain=1#L1991

hello, can I try this? newbie here, i don't know how to download and install PRs.
i think a lot of users would be super interested in using apollo's VDA on the fly virtual display with wgc as a service being the only way to access game bar and achievements notifications
tried  Vibepollo but it doesn't initialize sudovda virtual display in wgc if no real display is connected

@Nonary @ClassicOldSong

Issue #1200, opened 2025-11-03
getting a 401 when trying to hit /api/apps


AmanBhardwaj25:
Basically title. I am trying to setup some automation with moonlight/apollo

I am using the same command from LizardByte's documentation. 

`Invoke-RestMethod `
  -SkipCertificateCheck `
  -ContentType 'application/json' `
  -Uri 'https://localhost:47990/api/apps' `
  -Method GET `
  -Headers @{Authorization = 'Basic ' + [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes('user:pass'))}`

i get 401 error

`{
	"error": "Unauthorized",
	"status": false,
	"status_code": 401
}`

according to the docs it is the same admin username and password that we setup in the Web UI . I am able to log in just fine using that admin username password in the web ui. but using the same creds for the rest api i get a 401.

its responding with a 401 so it means API is definitely up. also /apps/configLocale returns a 200 too. 

i already tried resetting the username and password using `sunshine.exe --creds "newuser" "newpass"`
after resetting i can login to web ui still but i still get a 401 when trying to hit the REST API endpoints

Any idea how to solve this issue? I didnt find anything in the doc or online

ClassicOldSong:
Apollo uses cookies instead of basic auth. You need to post to the login API to get cookies first, then do other requests with the cookies.

AmanBhardwaj25:
> Apollo uses cookies instead of basic auth. You need to post to the login API to get cookies first, then do other requests with the cookies.

Awesome! thanks for pointing me in the right direction. i created a PR for updating documentation around auth and a couple endpoints i tested .

https://github.com/ClassicOldSong/Apollo/pull/1201

please take a look if you are interested

PR #1201, opened 2025-11-03
updated API documentation by adding info about accessing auth endpoint an…

AmanBhardwaj25:
…d apps endpoints.

This is just a documentation update for the endpoints i ended up testing. I noticed that the documentation still reflected what OG sunshine has. 

Based on what @ClassicOldSong  suggested under [https://github.com/ClassicOldSong/Apollo/issues/1200](url) this issue, i tested the auth endpoint and the GET and POST apps endpoints and updated the usage documentation 

I did not test any others as i have not needed to for my automation just yet.  when i do though i will create another PR

My automation im working on uses powershell so the example usages are based on that

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,19 +1,81 @@
 # API
 
-Sunshine has a RESTful API which can be used to interact with the service.
+Sunshine/Apollo has a RESTful API which can be used to interact with the service.
 
-Unless otherwise specified, authentication is required for all API calls. You can authenticate using
-basic authentication with the admin username and password.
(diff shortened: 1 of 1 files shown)


ClassicOldSong:
Thank you, but the docs are not actively maintained at this moment since they may change rapidly. Any upstream updates may cause the docs to be rewritten.

AmanBhardwaj25:
> Thank you, but the docs are not actively maintained at this moment since they may change rapidly. Any upstream updates may cause the docs to be rewritten.

sounds good.

PR #1256, opened 2025-11-24
Fix auto bitrate max cap to use original client-requested bitrate

AlexWorland:
## Problem

When auto bitrate is enabled and the user sets max bitrate to 500 Mbps in moonlight-qt, the actual bitrate was capped at ~398 Mbps instead of the requested 500 Mbps.

## Root Cause

The host applies several adjustments to the client-requested bitrate:
1. **FEC adjustment**: Reduces bitrate to account for FEC overhead
2. **Audio bitrate**: Subtracts audio bandwidth (~512 kbps for high-quality stereo)
3. **Overhead**: Subtracts 500 kbps for A/V packet overhead

These adjustments reduce 500 Mbps to ~398 Mbps. The auto bitrate controller was incorrectly using this **adjusted** bitrate as the maximum cap instead of the **original** client-requested bitrate.

## Solution

This PR stores the original client-requested bitrate (before FEC/audio/overhead adjustments) and uses it as the maximum cap for auto bitrate. This allows auto bitrate to reach the full client-requested maximum (500 Mbps in this case) while still applying the necessary adjustments to the initial encoding bitrate.

## Changes

- Added `original_client_bitrate_kbps` field to `launch_session_t` and `session_t`
- Store original bitrate in `rtsp.cpp` before adjustments are applied
- Copy original bitrate from launch session to stream session in `stream.cpp`
- Updated `auto_bitrate.cpp` to use original client bitrate as max instead of adjusted bitrate

## Testing

- Set max bitrate to 500 Mbps in moonlight-qt with auto bitrate enabled
- Verify that auto bitrate can now reach up to 500 Mbps (subject to network conditions)
- Verify that initial encoding bitrate still correctly accounts for FEC/audio/overhead

### DIFF
--- a/src/auto_bitrate.cpp
+++ b/src/auto_bitrate.cpp
@@ -123,8 +123,13 @@ namespace stream {
 
     int max_bitrate = settings.max_kbps;
     if (max_bitrate <= 0) {
-      // Use client's requested max or config max_bitrate
-      max_bitrate = session->config.monitor.bitrate;
+      // Use original client-requested bitrate (before FEC/audio/overhead adjustments) if available
+      // Otherwise fall back to adjusted bitrate
+      if (session->original_client_bitrate_kbps > 0) {
+        max_bitrate = session->original_client_bitrate_kbps;
+      } else {
+        max_bitrate = session->config.monitor.bitrate;
+      }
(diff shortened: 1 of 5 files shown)


AlexWorland:
Closing - PR should be against alexworland/apollo fork instead

PR #1286, opened 2025-12-16
Fix crash when boolean fields in sunshine_state.json are stored as strings

LTe:
Replace Boost property_tree with nlohmann::json in save_user_creds().

Boost property_tree does not preserve JSON types - it converts all values
to strings when writing JSON. This caused boolean and integer fields in
sunshine_state.json to become strings (e.g., true -> "true", 119480064 ->
"119480064") after changing the password, leading to crashes on restart.

nlohmann::json preserves the original types, matching the rest of the
codebase which already uses it for JSON operations.

### DIFF
--- a/src/httpcommon.cpp
+++ b/src/httpcommon.cpp
@@ -6,6 +6,7 @@
 
 // standard includes
 #include <filesystem>
+#include <fstream>
 #include <utility>
 
 // lib includes
@@ -68,11 +69,12 @@ namespace http {
   }
 
   int save_user_creds(const std::string &file, const std::string &username, const std::string &password, bool run_our_mouth) {
-    pt::ptree outputTree;
+    nlohmann::json outputTree;
 
     if (fs::exists(file)) {
       try {
-        pt::read_json(file, outputTree);
+        std::ifstream in(file);
+        in >> outputTree;
       } catch (std::exception &e) {
         BOOST_LOG(error) << "Couldn't read user credentials: "sv << e.what();
         return -1;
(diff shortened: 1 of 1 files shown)


ClassicOldSong:
Weird, I have already changed the whole settings parsing to proper JSON handling and changing these fields never got me into problems. Were you upgraded from a very old version or have reverted to an old version?

You should have noticed the format migration code and these kind of things should be handled there instead.

LTe:
@ClassicOldSong I can easily reproduce this on the fresh installation of version `0.4.6`.

1. Fresh install
2. Set initial password
3. Pair first device
4. Change Language
5. Save and Apply
6. Login again
7. Change password
8. Login again
9. Change language
10. Save and apply
11. Apollo is not starting

After this operation content of `C:\Program Files\Apollo\config\sunshine_state.json`

```json
{
    "password": "[REDACTED]",
    "root": {
        "named_devices": [
            {
                "allow_client_commands": "true",
                "always_use_virtual_display": "false",
                "cert": "[REDACTED]",
                "display_mode": "",
                "enable_legacy_ordering": "true",
                "name": "Mac",
                "perm": "119480064",
                "uuid": "9D2B43B6-AACC-510E-C417-029F89C51463"
            }
        ],
        "uniqueid": "C9E57A7A-04A9-ED0B-F71E-BAF1C6E09943"
    },
    "salt": "HM5xwNFijUhDEp2o",
    "username": "apollo"
}
```

LTe:
On the other hand I think `boost::property_tree` library does not output typed values in JSON and we always end up with string. So even when other parts of the code created boolean values it will just drop all the types and replace with string during password change.

So https://github.com/ClassicOldSong/Apollo/blob/fd059e36f082829a5eceeda7393df8dfb8c74073/src/httpcommon.cpp#L87 will cause issues.

ClassicOldSong:
I think I have replaced all the occurrences of boost property tree...

Anyway thanks for finding this out!

PR #1336, opened 2026-01-13
feat: Add client-side cursor support

TMCYTB:
# Adds client-side cursor support (transparent host cursor)

## Problem

When streaming over high-latency connections (e.g., 150ms+), mouse input feels extremely sluggish because:
1. User moves mouse on client
2. Input is sent to host (~150ms)
3. Host moves cursor and renders frame
4. Frame is encoded and sent back to client (~150ms)
5. **Total: ~300ms delay before seeing cursor movement**

This makes the streaming experience nearly unusable for productivity work over long distances.

## Solution

This PR adds support for **client-side cursor rendering**:
- When enabled, the host makes its cursor **transparent** during the streaming session
- The client renders its own local cursor that responds instantly to mouse movement
- Mouse input is still sent to the host normally, so clicks work at the correct position
- On session end (or crash), the host cursor is automatically restored

## How it works

1. Client sends `clientSideCursor=1` parameter when launching a session
2. Host receives this and calls `client_side_cursor_session_start(true)`
3. Host saves all system cursors (OCR_NORMAL, OCR_IBEAM, OCR_WAIT, etc.)
4. Host replaces them with a 1x1 transparent cursor
5. On session end, host restores all original cursors

## Safety features

- **Crash recovery**: Uses `SetUnhandledExceptionFilter` to restore cursors if Apollo crashes
- **Session counting**: Supports multiple concurrent sessions correctly
- **Emergency restore**: `SystemParametersInfoW(SPI_SETCURSORS)` as nuclear fallback

## Files changed

- `src/rtsp.h` - Added `clientSideCursor` field to `launch_session_t`
- `src/nvhttp.cpp` - Parse `clientSideCursor` parameter from client
- `src/stream.cpp` - Call cursor functions on session start/end
- `src/platform/common.h` - Declare platform functions
- `src/platform/windows/misc.cpp` - Windows implementation
- `src/platform/linux/misc.cpp` - Stub (Linux support could be added later)
- `src/platform/macos/misc.mm` - Stub (macOS support could be added later)

## Companion PR

This feature requires a corresponding client-side PR to moonlight-qt (https://github.com/moonlight-stream/moonlight-qt/pull/1785) that:
- Adds the "Client-side cursor" setting
- Sends the `clientSideCursor` parameter to the host
- Shows the local cursor during streaming (uses absolute mouse mode)

## Testing

Tested on Windows 11 host with ~150ms latency to client. Mouse feels instant while maintaining click accuracy, no more sloppy movements.

### DIFF
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -461,6 +461,7 @@ namespace nvhttp {
     launch_session->enable_hdr = util::from_view(get_arg(args, "hdrMode", "0"));
     launch_session->virtual_display = util::from_view(get_arg(args, "virtualDisplay", "0")) || named_cert_p->always_use_virtual_display;
     launch_session->scale_factor = util::from_view(get_arg(args, "scaleFactor", "100"));
+    launch_session->clientSideCursor = util::from_view(get_arg(args, "clientSideCursor", "0"));
 
     launch_session->client_do_cmds = named_cert_p->do_cmds;
     launch_session->client_undo_cmds = named_cert_p->undo_cmds;
--- a/src/platform/common.h
+++ b/src/platform/common.h
@@ -618,6 +618,10 @@ namespace platf {
   void streaming_will_start();
   void streaming_will_stop();
 
+  // Client-side cursor support (Windows only, no-op on other platforms)
+  void client_side_cursor_session_start();
+  void client_side_cursor_session_stop();
+
   void restart();
 
   /**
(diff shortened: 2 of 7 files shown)


ClassicOldSong:
You can use Ctrl+Shift+Alt+N to disable host side cursor capture directly. There's no need for a transparent cursor.

Also, a proper solution should still capture the cursor but just don't render it in stream, instead use a separate channel to send the captured cursor and render it on the client side.

PR #1348, opened 2026-01-18
Fix headless mode: Set virtual display as primary automatically

ky0k1ng:
Fixes issue where applications (like Steam) launch on the physical monitor instead of the virtual display when headless mode is enabled.

Problem
When `headless_mode` is enabled and a virtual display is created, Steam and other applications still launch on the physical monitor because the virtual display is not set as primary.

Solution
- Automatically configure virtual display as primary when `headless_mode` is enabled
- Uses existing `display_device::configure_display()` API with `ensure_primary` flag
- Temporarily overrides `dd.configuration_option` during setup
- Properly manages display topology through SettingsManager instead of direct Win32 calls

Testing
Tested on Windows with NVIDIA RTX 3080 Ti and SudoVDA driver:
- Applications (Steam) now correctly launch on virtual display in headless mode
- Display configuration properly reverts after session endsoach uses the existing SettingsManager API instead of direct Win32 calls for more reliable display management.

### DIFF
--- a/src/process.cpp
+++ b/src/process.cpp
@@ -325,6 +325,29 @@ namespace proc {
           // empty name when probing graphics cards.
 
           config::video.output_name = display_device::map_display_name(this->display_name);
+
+          // In headless mode, ensure virtual display becomes primary
+          // We need to temporarily override configuration_option to ensure_primary
+          if (config::video.headless_mode) {
+            BOOST_LOG(info) << "Headless mode: Configuring virtual display as primary";
+            
+            // Save original configuration option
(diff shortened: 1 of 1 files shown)


ClassicOldSong:
Thank you for the PR but this is not the designed behavior. You're supposed to set the virtual display as primary manually in Windows settings. Windows will remember the config automatically. This leaves freedom for users to choose whether they want the virtual display to be primary, secondary, or mirror.

Lots of people use Apollo for extended display so don't force a config. Also, I have made clear in a warning saying that DisplayDeviceManager actually causes inconsistency problems, so I don't recommend using it.

PR #1375, opened 2026-02-02
Send external_ip config property to clients

Unnvaldr:
### Rationale
There is no mechanism that determines the external IP address of the machine, which is crucial especially for computers behind VPN/NAT. Sending the manually set external IP address to the client, will allow for the machine to be discoverable in local and remote scenarios.
### What Changed
Addition of an additional JSON property `root.ExternalIP` to the Nvidia GameStream's protocol, that reflects the server's config `external_ip` property value.
### Impact
No change in behaviour on the server's side; clients will have one additonal endpoint to check at discovery and to send to for WoL packets.

### DIFF
--- a/src/nvhttp.cpp
+++ b/src/nvhttp.cpp
@@ -958,6 +958,8 @@ namespace nvhttp {
       tree.put("root.LocalIP", net::addr_to_normalized_string(local_endpoint.address()));
     }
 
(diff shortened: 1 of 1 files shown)


ClassicOldSong:
Are there clients supporting this field right now?

Unnvaldr:
Moonlight based clients have usages for this external IP, but by default it is empty and as a result, ignored.

PR #1418, opened 2026-03-10
Add seat abstraction and per-seat video capture

ccowan93:
## Summary
- **Phase 1 - Seat Abstraction Layer**: Introduces `seat_t` struct bundling display, audio, and input resources per session. `seat::manager` handles thread-safe acquire/release. Integrated into RTSP session setup and stream lifecycle.
- **Phase 2 - Per-Seat Video Capture**: Replaces global `capture_thread_async` with a per-display capture thread map. Each seat can target its own display independently. Fully backward-compatible in single-seat mode.
- Fixes uninitialized `_count` in `safe::shared_t` for heap-allocated instances.

## Test plan
- [ ] Verify single-seat streaming still works identically (backward compatibility)
- [ ] Confirm seat is acquired on RTSP ANNOUNCE and released on session end
- [ ] Test that session start failure properly releases the seat
- [ ] Verify capture thread is created per unique display name
- [ ] Check that `proc::proc.display_name` is only updated in single-seat mode

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmake/compile_definitions/common.cmake
+++ b/cmake/compile_definitions/common.cmake
@@ -87,6 +87,8 @@ set(SUNSHINE_TARGET_FILES
         "${CMAKE_SOURCE_DIR}/src/confighttp.h"
         "${CMAKE_SOURCE_DIR}/src/rtsp.cpp"
         "${CMAKE_SOURCE_DIR}/src/rtsp.h"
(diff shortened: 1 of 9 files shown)


ClassicOldSong:
Kind suggestion: don't waste your precious Claude tokens on this hot mess. I'll solve this problem in another way in the future.

PR #1460, opened 2026-04-14
fix(audio/windows): set waveformat when needed

HarmfulBreeze:
The waveformat should not be set on every client connection, as it could stop the audio sessions for some games (like Slay the Spire).

This PR addresses this issue by adding a simple cache map meant to remember device_id <-> waveformat associations, and only changing the waveformat *if needed*.

### DIFF
--- a/src/platform/windows/audio.cpp
+++ b/src/platform/windows/audio.cpp
@@ -56,6 +56,8 @@ namespace {
                                               SPEAKER_BACK_LEFT | SPEAKER_BACK_RIGHT |
                                               SPEAKER_SIDE_LEFT | SPEAKER_SIDE_RIGHT;
 
+  std::map<std::wstring, WAVEFORMATEX> cached_device_formats = {};
+
   enum class sample_format_e {
     f32,
     s32,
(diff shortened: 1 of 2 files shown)


HarmfulBreeze:
@ClassicOldSong friendly bump :)

src/platform/windows/audio.cpp
```diff
@@ -345,15 +347,13 @@ namespace platf::audio {
     );
 
     if (FAILED(status)) {
-      BOOST_LOG(error) << "Couldn't get default audio endpoint [0x"sv << util::hex(status).to_string_view() << ']';
```
ClassicOldSong, line 348: Why removing this log?

ClassicOldSong:
Sorry I was really busy recently and I forgot to take a look afterwards...

HarmfulBreeze:
It's fine, no worries. :)

Thanks for your feedback. I reverted the error logging in `default_device`, I think it made sense to me to remove it at the time, as I wanted to make the caller check for it since it now returns an `std::optional`. But I never went ahead and lifted the error handling so... yeah

PR #1465, opened 2026-04-15
feature(audio): release audio sink on client pause

RJNY:
I was going crazy trying to figure out why the PC refused to sleep ([thread](https://gist.github.com/ClassicOldSong/0a088d2a92a06f86e48fe328b67ec995?permalink_comment_id=5930788#gistcomment-5930788)) while using your AHK sleep scripts and realized it's one of 2 things:
- game full screen prevents sleep. (obviously, but irrelevant to this PR)
- audio sink holds audio driver hostage on disconnect (directly relates to this PR)

This change introduces a fix for the second bullet point. It releases the audio driver when all clients disconnect, so it no longer prevents the PC from sleeping.
I've tested these changes on my local machine for over a week now using my own personal [AHK script](https://gist.github.com/RJNY/a1eaf1ca640ae654401f3995221f5a5a) (still needs work) that freezes and minimizes the game, allowing the host PC to sleep on disconnect!

Trying not to go too deep into the AHK script specifics on this audio sink PR, but thought it was relevant to understanding _**why**_ I want this change.

This is an opt-in feature. original behavior is preserved by default.

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -308,18 +308,40 @@ namespace audio {
     return 0;
   }
 
+  static void revert_audio_sink(audio_ctx_t &ctx) {
+    // Change back to the host sink, unless there was none
+    const std::string &sink = ctx.sink.host.empty() ? config::audio.sink : ctx.sink.host;
+    if (!sink.empty()) {
+      BOOST_LOG(info) << "Restoring audio sink to: "sv << sink;
+      // Best effort, it's allowed to fail
+      ctx.control->set_sink(sink);
+    }
+  }
+
   void stop_audio_control(audio_ctx_t &ctx) {
     // restore audio-sink if applicable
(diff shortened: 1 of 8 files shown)


ClassicOldSong:
Thank you but why change the title? Technically you should only release it when paused, which means all clients are disconnected.

RJNY:
> Thank you but why change the title? Technically you should only release it when paused, which means all clients are disconnected.

the short answer is, I was arguing with myself about semantics. 

I was concerned about conflating the two concepts. For my use case, I am “pausing” with pause.ahk, but likely a majority of users don’t use the ahk scripts to pause, they disconnect but keep the session alive. So I was making sure not to conflate the two and keep the language neutral and not break expected behavior without scripts. but the process is indeed “paused”, so I had it right the first time. 

reverted title back to pause.

PR #1491, opened 2026-05-14
Fix crash when paired-client booleans in sunshine_state.json are stored as strings

AntonKun:
## Problem

On Windows 11, `Sunshine.exe` (Apollo v0.4.6) could crash on startup with an unhandled nlohmann::json exception:

`[json.exception.type_error.302] type must be boolean, but is string`

In the Windows Event Log this often showed up as a fault in `ucrtbase.dll` with exception code `0xc0000409` (abort after an unhandled exception / `std::terminate`).

In `sunshine_state.json`, under `root.named_devices`, the fields `allow_client_commands`, `always_use_virtual_display`, and `enable_legacy_ordering` were sometimes persisted as JSON strings (e.g. `"true"`) instead of JSON booleans. Reading them with `nlohmann::json::value(key, default_bool)` requires an actual boolean type, so parsing threw before initialization completed -sometimes leading to a service restart loop.

## Solution

Use the existing helper `util::get_non_string_json_value<bool>` for these three fields (same approach as for the `elevated` flag in client command entries), so both JSON booleans and string values like `"true"` are accepted.

- `src/nvhttp.cpp` - loading `file_state` (`sunshine_state.json`).
- `src/confighttp.cpp` - client update API, so the same string-encoded booleans in the request body do not trigger the same exception.

### DIFF
--- a/src/confighttp.cpp
+++ b/src/confighttp.cpp
@@ -909,9 +909,9 @@ namespace confighttp {
       std::string uuid = input_tree.value("uuid", "");
       std::string name = input_tree.value("name", "");
       std::string display_mode = input_tree.value("display_mode", "");
-      bool enable_legacy_ordering = input_tree.value("enable_legacy_ordering", true);
-      bool allow_client_commands = input_tree.value("allow_client_commands", true);
-      bool always_use_virtual_display = input_tree.value("always_use_virtual_display", false);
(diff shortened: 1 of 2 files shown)


ClassicOldSong:
Thanks but a better fix was already merged, just not released.

https://github.com/ClassicOldSong/Apollo/pull/1286

PR #1492, opened 2026-05-17
Linux Ubuntu 26.04 Support with Virtual Display via EVDI

primez-x:
Creating PR for visibility purposes mostly.

**What This Fork Provides**
- GNOME Wayland virtual display streaming for Ubuntu.
- EVDI-backed real virtual monitors so streamed sessions do not mirror the physical display.
- Mutter ScreenCast/PipeWire capture for smooth frame pacing on GNOME.
- AMD, Intel, and Nvidia encoding through the Linux encoder stack available on the host.
- User service packaging, udev rules, EVDI module loading, and Ubuntu install documentation.

**Supported Host**
- Ubuntu 24.04 LTS or newer.
- Ubuntu 26.04 development/current testing builds.
- GNOME on Wayland with the streaming user already logged in.
- A Moonlight-compatible client such as Artemis or Moonlight.
- Other Linux distributions may still build, but Ubuntu is the supported release target for this fork.


ClassicOldSong:
Thank you but I think EVDI isn't really suitable for this job here as it can't utilize vram directly. Your frame buffer memory is on your system memory not your GPU, making things much slower and wasting PCIE bandwidth. Correct me if I'm wrong.

There's another PR that does things on Linux with a much smarter way and doesn't require installing an unsigned driver.

For a PR that contains thousands of changes with conflicting implementation to an already existing PR, I'm confused, and this kind of fragmentation is what exactly what I wanted to avoid in the feature bounty RFC (checkout discussions). The fact that Linux is so fragmented is exactly why I didn't provide Linux build of Apollo in the first place...

Maybe Apollo will supports macOS earlier than Linux officially, since at least, macOS has a single source of truth for doing virtual displays and it's already practiced in many other separated projects.

ClassicOldSong:
Also I think you know evdi is a Linux kernel module which doesn't have anything to do with a distro, specifying Ununtu here makes me think you're not actually familiar with Linux enough. Still, correct me if I'm wrong.

IrakliChkhetia:
>There's another PR that does things on Linux with a much smarter way and doesn't require installing an unsigned driver.

Correct me if I’m wrong, but using EDID overrides via debugfs requires the GPU to have at least one free port available. What should someone do in that case if they already have 4 monitors connected?

primez-x:
> Also I think you know evdi is a Linux kernel module which doesn't have anything to do with a distro, specifying Ununtu here makes me think you're not actually familiar with Linux enough. Still, correct me if I'm wrong.

I am not an expert in Linux at all. I specified Ubuntu primarily because I had to create this fork for 26.04 Ubuntu compatibility. I don’t recall exactly but something in 26.04 was different that required quite a bit of rework. Primary reason for me creating the PR was not to merge code, mostly just to share what is working.

The ACTUAL parent for this fork is actually the Linux port from ~2 years ago from some other guy who did virtual displays via EVDI. I didn’t see any virtual display support from the native Apollo fork, so I started from his (hence why it’s a massive amount of code change most likely).

That said, this fork uses EVDI for simulated the monitor but everything else is done in VRAM. EVDI capture actually didn’t work well at all in my experience. Far too slow and unresponsive.

Took a lot of trial/error to get it working, but the core concept applies so I wanted to share the fork in concept so that perhaps someone could use something here if needed.

And other persons comment is right. EVDI does not require a port open as it’s not emulating a physical monitor but actually creating a new virtual monitor entirely. 

This fork probably could be cleaned up dramatically by starting fresh from your fork then keeping only the good / working components for EVDI and Wayland capture.

ClassicOldSong:
IIRC EVDI isn't actually using your vram for display frame buffers: https://displaylink.github.io/evdi/quickstart/#frame-buffers

primez-x:
This is valid. Let me explore some other options. EVDI's capture is CPU bound, which is why the performance was so terrible in my early testing of EVDI. EVDI is only the virtual monitor provider in my fork at the moment. The actual stream capture is routed through GNOME Mutter ScreenCast/PipeWire, then through the normal Linux encoder path -- but there probably is still some EVDI overhead...

So this is a hybrid approach, but I guess that also means there is still some work for me to do here to be 100% GPU sourced.

PR #1496, opened 2026-05-21
[codex] Backport client certificate validation fix

WatchKitty:
## Summary

Backports the upstream Sunshine client certificate validation fix for GHSA-ph75-mgxh-mv57 / CVE-2026-32253.

This removes the special-case success path for `X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY`, preventing an untrusted client certificate chain from being accepted as authorized.

## Provenance

Cherry-picked from LizardByte/Sunshine commit `888a6bb0cbdab97afb3df712f85f19c09dccd367` using `git cherry-pick -x`, preserving the original author metadata.

Upstream patched release: `v2026.516.143833`.

## Validation

- Confirmed `src/crypto.cpp` no longer contains `X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY` in the OpenSSL verify callback.
- Did not run a full Windows build locally.

### DIFF
--- a/src/crypto.cpp
+++ b/src/crypto.cpp
@@ -33,8 +33,6 @@ namespace crypto {
       // Expired or not-yet-valid certificates are fine. Sometimes Moonlight is running on embedded devices
       // that don't have accurate clocks (or haven't yet synchronized by the time Moonlight first runs).
       // This behavior also matches what GeForce Experience does.
-      // TODO: Checking for X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY is a temporary workaround to get moonlight-embedded to work on the raspberry pi
-      case X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY:
       case X509_V_ERR_CERT_NOT_YET_VALID:
       case X509_V_ERR_CERT_HAS_EXPIRED:
         return 1;


WatchKitty:
@ClassicOldSong Since this PR has been merged and it backports the GHSA-ph75-mgxh-mv57 / CVE-2026-32253 client certificate validation fix, could you please publish a new stable Apollo release when you have a chance?

The current latest stable GitHub Release is still v0.4.6, and v0.4.7-alpha.1 is marked as a pre-release. There is also a v0.4.8 tag, but no corresponding GitHub Release asset, and it appears to predate this security fix.

A release based on adc5c5a0bd80831ce495434bb16aee2cd4175fb8 or current master would make it much easier for regular users to get the fix without building from source. Thank you!

ClassicOldSong:
There needs much more work to do, the last pre-release was removed due to a regression in ffmpeg causing issues on AMD cards, now it's still on that commit, at least that part needs a bump

gfreitash:
Given how critical this port is for web-exposed instances, this should really be cherry-picked into the last release tag and be released as a hotfix

Gevroska:
I agree with @gfreitash here. Would it be possible to release a minimal 0.4.6 security hotfix containing just the certificate validation fix from adc5c5a, instead of waiting for the next full release?
I checked my installed v0.4.6 sunshine.exe and confirmed that the vulnerable X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY success path is actually present in the compiled binary, so this isn't only theoretical/source-level.
For users exposing Apollo's GameStream ports to the Internet, especially TCP 47984, this leaves a known critical authentication bypass reachable remotely unless they close the port, use a VPN, or build Apollo themselves.
I understand that the next release is currently blocked by the FFmpeg/AMD regression, but that seems separate from this security fix. A small hotfix based on the stable release with only #1496 cherry-picked would be really useful.
Thanks!

ClassicOldSong:
The simplest solution is DO NOT directly expose Apollo on the internet. It's not wise to do so even with this issue fixed.

Gevroska:
First of all, thank you very much for all the work you put into Apollo. I know this is a volunteer project, and I genuinely appreciate it a lot. I use Apollo every day, and for me it provides a lot of value compared to Sunshine.
I do understand and agree with the recommendation not to expose Apollo directly to the Internet, and I agree that using a VPN is the safer setup. Even with this issue fixed, there could always be another unknown vulnerability.
That said, I think this is slightly separate from whether the known vulnerability should still be fixed in a release. Apollo, like Sunshine, does provide UPnP support and therefore makes direct Internet exposure an available/possible configuration. Some users are already using it that way and will probably continue to do so, even if it is not the recommended setup.
So from my perspective, it would still make sense to provide a build containing the already-merged fix, even though you have repeatedly stated that direct Internet exposure is discouraged and that a VPN is recommended.

PR #1497, opened 2026-05-21
Add GitHub Actions release workflow and fix Windows build

WatchKitty:
## Summary

Adds a GitHub Actions release workflow that can build the Windows NSIS installer on GitHub-hosted Windows runners.

The workflow supports:

- pull request builds that upload the installer as a workflow artifact
- tag pushes matching `v*` that build and publish a GitHub Release asset
- manual `workflow_dispatch` releases with configurable tag, draft, prerelease, and publish toggles
- reruns against an existing release by updating release metadata and re-uploading the installer asset with `--clobber`

This also fixes a current Windows build blocker in the AMF H.264 options: `config_t` has no `profile` field, so the AMF encoder now sets the existing fallback value (`high`) directly.

## Why

The repository already documents remote builds, and the recent GHSA-ph75-mgxh-mv57 / CVE-2026-32253 backport would be easier for regular users to consume if maintainers can produce a cloud-built installer and release asset directly from GitHub.

## Validation

- Parsed `.github/workflows/release.yml` with `yaml@2`.
- Ran `actionlint` against the workflow.
- Let the workflow run on this PR; the first run exposed missing `npm` in the MSYS2 shell, then the second run exposed the existing `config_t::profile` Windows build error. Both have been addressed in this branch.


ClassicOldSong:
The workflow was deliberately removed since it's way too chaos

ClassicOldSong:
Also, due to the term "CI", each push into a PR will get built and due to how people generally want to chase "new", malicious commits can use this as an attack surface, so it's better left off.

WatchKitty:
Thanks, that makes sense. I didn’t realize the release workflow had been intentionally removed before.

PR #1594, opened 2026-09-11
Add configurable WAN video pacing

Swanand01:
Adds optional RTP video pacing and batch-size controls for WAN testing while preserving legacy behavior by default.

Also adds debug telemetry for sender pacing, packet queue wait, batch size, and packets per frame.


Swanand01:
Closing: this was opened against upstream by mistake. The branch will remain on the fork for private testing.

